PR #7, opened 2020-09-10
[Python] - Add support for start/end NVTX ranges

DEKHTIARJonathan:


python/nvtx/nvtx.py
```diff
@@ -149,3 +153,46 @@ def pop_range(domain=None):
...
+
+    Examples
+    --------
+    >>> import time
+    >>> import nvtx
+    >>> marker_id = nvtx.start_range("my_code_range", domain="my_domain")
+    >>> time.sleep(1)
+    >>> nvtx.end_range(marker_id, domain="my_domain")
+    """
+    return libnvtx_start_range(
```
shwina: Rather than returning an `int` here, should we return the opaque object of type `RangeId` instead? Similarly for the `end_range` function below.

DEKHTIARJonathan: I did this because we re-create the "Domain" object from string. Instead of returning it. I don't see why they should be treated differently.

And this follows how you implemented push/pop range

DEKHTIARJonathan (commented):
Comment fixed

python/nvtx/nvtx.py
```diff
@@ -149,3 +153,48 @@ def pop_range(domain=None):
...
+    ----------
+    message : str, optional
+        A message associated with the annotated code range.
+    color : str, color, optional
+        A color associated with the annotated code range.
+        Supports
+    domain : str, optional
+        Name of a domain under which the code range is scoped.
+        The default domain name is "NVTX".
+
```
shwina, line 172:
```suggestion

Returns
--------
id : RangeId
    An object that must be passed to the `end_range` function to mark the end of the annotated code range.

```

shwina:
Giving this some additional thought, I think it would be better to hold a reference to the `Domain` object in the `RangeId` object. This simplifies usage from the end user standpoint a bit:

So rather than:

```python
rng_id = nvtx.start_range("my_code_range", domain="my_domain")
time.sleep(1)
nvtx.end_range(rng_id, domain="my_domain")
```

```python
rng_id = nvtx.start_range("my_code_range", domain="my_domain")
time.sleep(1)
nvtx.end_range(rng_id)
```

Thoughts? Apologies for the churn. :(

shwina:
Merging as is, and will open a follow-up PR with Cython changes. Thanks @DEKHTIARJonathan!

PR #8, opened 2020-09-15
[REVIEW] Improvements to start_range/end_range Python APIs

shwina:
cc: @kkraus14 

- [x] Implements the suggestion in https://github.com/NVIDIA/NVTX/pull/7#issuecomment-692041395
- [x] Adds docs for start_range/end_range
- [x] Adds `category` argument to all APIs

### DIFF
--- a/python/docs/basic.rst
+++ b/python/docs/basic.rst
@@ -33,6 +33,24 @@ name of the function being decorated:
        pass
 
 
+start_range and end_range
+-------------------------
+
+In certain situations, it is impossible to use ``annotate()``,
+e.g., when a code range spans multiple functions.
+In such cases, the ``start_range()`` and ``end_range()`` functions
+can be used instead.
+
+The ``start_range()`` function is called at the beginning of a code range,
+and returns a handle. The handle is passed to the ``end_range()`` function,
+which is called at the end of the code range.
+
+::
+   rng = nvtx.start_range(message="my_message", color="blue")
+   # ... do something ... #
+   nvtx.end_range(rng)
+
+
 mark
 ----
 
@@ -127,7 +145,6 @@ Although `func_1` and `func_3` are both grouped
 under a category named `Cat_1`, they are unrelated
 as each domain maintains its own categories.
 
-Unlike domains, caetgories are not expensive to create and manage.
+Unlike domains, categories are not expensive to create and manage.
 Thus, you should prefer categories for maintaining several groups
 of annotations.
-
--- a/python/docs/basic.rst
+++ b/python/docs/basic.rst
@@ -37,7 +37,7 @@ start_range and end_range
 -------------------------
 
 In certain situations, it is impossible to use ``annotate()``,
-e.g., when a code range spans multiple functions.
+e.g., when a code range spans multiple functions or in asynchronous code.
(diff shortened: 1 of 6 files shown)


python/docs/basic.rst
```diff
@@ -33,6 +33,24 @@ name of the function being decorated:
        pass
 
 
+start_range and end_range
+-------------------------
+
+In certain situations, it is impossible to use ``annotate()``,
```
kkraus14, line 39: Maybe allude to async Python functions as well?

python/nvtx/nvtx.py
```diff
@@ -111,8 +110,13 @@ def mark(message=None, color="blue", domain=None):
         the event is scoped. If unspecified, the event is not associated
         with a category.
     """
-    attributes = EventAttributes(message, color)
     domain = Domain(domain)
+    category_id = (
+        self.domain.get_category_id(category)
+        if category is not None
+        else None
```
kkraus14, line 117: Should we have `get_category_id` handle `None` instead of having to protect it everywhere?

python/nvtx/tests/test_basic.py
```diff
@@ -153,3 +153,48 @@ def test_get_category_id():
...
+    [
+        None,
+        "",
+        "y"
+        "x",
+        "abc",
+        "abc def"
+    ]
+)
+def test_start_end(message, color, domain, category):
```
kkraus14, line 186: This generates 500 different test permutations, are all of these necessary?

python/nvtx/nvtx.py
shwina: That's a good suggestion, but I like this better (even if it is repetitive and ugly) because it's more explicit about how `None` is handled. If you feel strongly, I'll change.

python/nvtx/tests/test_basic.py
shwina: Not really -- reduced :)

python/docs/basic.rst
shwina: How does the `start/end` API help with async functions? (not contesting your point, but actually curious)

python/nvtx/nvtx.py
kkraus14: Nope, just a question, defer to your judgement here.

python/docs/basic.rst
kkraus14: Async functions make no guarantee about when they end when started so it's easy to have a pattern that looks like:

- start A
- start B
- end A
- start D
- end C
- end D
- end B

Which makes using a stack based approach of push/pop impossible.

shwina: Gotcha -- yeah that makes sense.

python/nvtx/nvtx.py
shwina: Let's keep this for now

python/docs/basic.rst
shwina: Added async code. I realized we don't document the `push_range` and `pop_range` APIs because there's no really good reason to use them, as compared to `annotate`.

kkraus14 (approved):

PR #17, opened 2021-02-04
Fix handling category in push/pop and mark APIS

### LINKED ISSUE #16: Python nvtx.push_range function missing category argument
https://github.com/NVIDIA/NVTX/blob/24a123e8b2c6bd498a27dbe867a96a8c01b0f583/python/nvtx/nvtx.py#L123

this should be

```py
def push_range(message=None, color="blue", domain=None, category=None):
```

since `category` is used here.

https://github.com/NVIDIA/NVTX/blob/24a123e8b2c6bd498a27dbe867a96a8c01b0f583/python/nvtx/nvtx.py#L152

shwina:
Fixes #16

### DIFF
--- a/python/nvtx/nvtx.py
+++ b/python/nvtx/nvtx.py
@@ -92,7 +92,7 @@ class annotate(ContextDecorator):
         return super().__call__(func)
 
 
-def mark(message=None, color="blue", domain=None):
+def mark(message=None, color="blue", domain=None, category=None):
     """
     Mark an instantaneous event.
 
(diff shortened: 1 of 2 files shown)


python/nvtx/nvtx.py
```diff
@@ -112,15 +112,15 @@ def mark(message=None, color="blue", domain=None):
     """
     domain = Domain(domain)
     category_id = (
-        self.domain.get_category_id(category)
+        domain.get_category_id(category)
```
shwina, line 115: this was so broken...

kkraus14 (approved):

PR #19, opened 2021-02-04
Bump to 0.2.3

shwina:


shwina:
I made a mistake and 0.2.2 wheels on PyPI are broken. Thus cutting a new release.

PR #22, opened 2021-02-18
[Python] - Add support for integer type category argument

beru:
This PR will add support for integer type category argument.

By allowing direct integer type for category argument, hoping that category id bridge between Python script and C/C++ code would be guaranteed.

I suspect choosing string type category argument is done in the name of abstraction and ease of use.
But some people crave for magic numbers, so the road for those poor souls must be paved with good intentions.

### DIFF
--- a/python/nvtx/nvtx.py
+++ b/python/nvtx/nvtx.py
@@ -40,10 +40,10 @@ class annotate(ContextDecorator):
         domain : str, optional
             A string specifying the domain under which the code range is
             scoped. The default domain is named "NVTX".
-        category : str, optional
-            A string specifying the category within the domain under which
-            the code range is scoped. If unspecified, the code range is not
-            associated with a category.
+        category : str, int, optional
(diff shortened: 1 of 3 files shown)


python/nvtx/nvtx.py
```diff
@@ -64,12 +64,12 @@ def __init__(self, message=None, color=None, domain=None, category=None):
...
-            self.domain.get_category_id(category)
-            if category is not None
-            else None
-        )
+ 
+        category_id = None
+        if isinstance(category, int):
+            category_id =  category
+        elif isinstance(category, str):
+            category_id = self.domain.get_category_id(category)
```
shwina, line 72: This logic seems to be repeated in a few different places -- should we just make it part of `get_category_id`?

shwina:
@beru thanks for your PR and apologies for not reviewing earlier.

jcohen-nvidia:
Sorry for the excessive delays, thanks @beru!

PR #32, opened 2021-03-23
Running NVTX as a module automatically annotates functions

shwina:
This PR adds the ability to the Python bindings for NVTX to automatically annotate _every_ function call (including, if desired, C functions).

This is done by running `nvtx` as a module. Example usage:

```
 # adds NVTX ranges around every Python call in program.py:
 nsys profile -t cuda,nvtx python -m nvtx program.py 
```


jcohen-nvidia:
Hi @shwina  -- I am finally getting back to NVTX and I noticed this PR has conflicts now.  Can you resolve these?  Then I can merge.  Thanks!

shwina:
Thanksk @jcohen-nvidia - this should be good now.

PR #39, opened 2021-03-27
cpp: fix error when `NVTX_DISABLE` is defined

### LINKED ISSUE #38: Compilation error when if `NVTX_DISABLE` is defined
The following code fails to compile with GCC 8.4.0.

```
#define NVTX_DISABLE
#include <nvtx3/nvtx3.hpp>
int main() {}
```

Compilation error:

```
/home/afroger/dev/projects/NVTX/cpp/include/nvtx3/nvtx3.hpp: In function ‘nvtx3::v1::range_handle nvtx3::v1::start_range(const Args& ...)’:
/home/afroger/dev/projects/NVTX/cpp/include/nvtx3/nvtx3.hpp:1979:9: error: ‘first’ was not declared in this scope
   (void)first;
         ^~~~~
```

AntoineFroger:
Closes #38

### DIFF
--- a/cpp/include/nvtx3/nvtx3.hpp
+++ b/cpp/include/nvtx3/nvtx3.hpp
@@ -1976,7 +1976,6 @@ range_handle start_range(Args const&... args) noexcept
 #ifndef NVTX_DISABLE
   return start_range(event_attributes{args...});
 #else
-  (void)first;
   return range_handle{};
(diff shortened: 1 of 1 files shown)


jrhemstad (approved):

jcohen-nvidia:
Good catch.

PR #49, opened 2022-01-26
Support Python 3.9 wheels and bump to v0.2.4

shwina:


python/tools/build-wheels.sh
```diff
@@ -6,15 +6,15 @@ function repair_wheel {
     if ! auditwheel show "$wheel"; then
         echo "Skipping non-platform wheel $wheel"
     else
-        auditwheel repair "$wheel" --plat "manylinux2010_x86_64" -w /io/python/wheels/
+        auditwheel repair "$wheel" --plat "manylinux2014_x86_64" -w /io/python/wheels/
```
jakirkham, line 9: Where does the build happen? Asking as the Docker image used to do the build would need to be updated as well. `manylinux2010` is CentOS 6 based whereas `manylinux2014` is CentOS 7 based

shwina: Thanks  --  currently the builds are happening locally. FWIW, I run Ubuntu 20.04, so the newer Docker image should be no problem.

python/setup.py
```diff
@@ -46,6 +46,7 @@
         "Programming Language :: Python :: 3.6",
         "Programming Language :: Python :: 3.7",
         "Programming Language :: Python :: 3.8",
+        "Programming Language :: Python :: 3.9",
     ],
     # Include the separately-compiled shared library
     setup_requires=["Cython"],
```
jakirkham: Is this still needed given we are using [`build-system` in `pyproject.toml`]( https://github.com/NVIDIA/NVTX/blob/3c98c8425b0376fd8653aac7cfc6a864f3897752/python/pyproject.toml#L21-L22 )?

python/tools/build-wheels.sh
jakirkham: Sorry was more confused by the fact this script is called `build-wheels.sh`, but doesn't seem to include a build step

Ok though the build is happening in a `manylinux` Docker image, right?

shwina: Correct. Specifically I was going to use  quay.io/pypa/manylinux2014_x86_64

python/setup.py
shwina: Fix'd

jakirkham (approved):

PR #60, opened 2022-10-06
NVTX3 requires CMake 3.19+ as we set properties on INTERFACE targets

### LINKED ISSUE #98: Meet cmake error when building nvtx3 on Jetson Orin
Hi!

I meet a cmake error when building nvtx3 on Jetson Orin, which is

```
CMake Error at imports/NVTX/nvtxImportedTargets.cmake:54 (set_target_properties):
  INTERFACE_LIBRARY targets may only have whitelisted properties.  The
  property "VERSION" is not allowed.
Call Stack (most recent call first):
  imports/NVTX/CMakeLists.txt:33 (include)


CMake Error at imports/NVTX/nvtxImportedTargets.cmake:64 (set_target_properties):
  INTERFACE_LIBRARY targets may only have whitelisted properties.  The
  property "VERSION" is not allowed.
Call Stack (most recent call first):
  imports/NVTX/CMakeLists.txt:33 (include)
```

The version of cmake is 3.16.3
By the way, I have sucessfully built nvtx3 on my laptop and the verison of cmake is 3.16.3

robertmaynard:
The CMake 3.19 relaxation on `INTERFACE` targets to have sources aslo removed all restrictions around which properties an INTERFACE target can have ( https://gitlab.kitware.com/cmake/cmake/-/merge_requests/5078 ). NVTX3 is adding the version property to INTERFACE targets and so requires 3.19+

### DIFF
--- a/c/CMakeLists.txt
+++ b/c/CMakeLists.txt
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #=============================================================================
-cmake_minimum_required(VERSION 3.10)
+cmake_minimum_required(VERSION 3.19)
 
 # This file is typically used by calling add_subdirectory on this directory or add
 # this directory as an external package.  In this case, the NVTX library targets
--- a/c/nvtxImportedTargets.cmake
+++ b/c/nvtxImportedTargets.cmake
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #=============================================================================
-cmake_minimum_required(VERSION 3.10)
+cmake_minimum_required(VERSION 3.19)
 
 set(NVTX3_VERSION 3.1.0)
 


invisiblepancake (changes requested):
Upon examining the modifications introduced in commit [a649f93c4cf3f690ab376081be303c01e2032ea6](https://github.com/jart/bestline/commit/a649f93c4cf3f690ab376081be303c01e2032ea6), I am compelled to convey my observations in the fashion of our venerable ancestors.

In the function bestlineEdit:

A new variable promptnotnull has been introduced to ensure the prompt is never null, defaulting to an empty string if necessary.
Another variable, promptlastnl, seeks out the final newline character within the prompt. This is a prudent measure, ensuring user prompts are handled with the utmost propriety.
The prompt used by bestlineWriteStr is now promptnotnull, which maintains the integrity of the original prompt.
Alteration in bestlineInit:

The condition has been refined to exclude newline characters from the validity check of the prompt. This change aligns with contemporary practices, ensuring the prompt's format adheres to the expected standards.
These refinements exhibit a meticulous attention to detail, reinforcing the robustness of our beloved code. May these changes stand the test of time, much like the enduring works of our forebears.

Yours in the pursuit of excellence,
[Your Name]

PR #64, opened 2022-11-21
remove hidden pragmas

simonbyrne:
These prevent building of a shared library (as described in https://github.com/NVIDIA/NVTX#other-languages), and appear to be redundant in any case.

See https://forums.developer.nvidia.com/t/building-nvtx-as-a-dynamic-library/235031?u=simonbyrne1


evanramos-nvidia:
A fix for `#define NVTX_EXPORT_API` as you mentioned in your forum post is pending for the upcoming patch release.

PR #67, opened 2023-03-06
Add support for `NVTX_DIR`, drop Python 3.7

shwina:


python/setup.py
```diff
@@ -23,6 +23,9 @@
 include_dirs = [os.path.dirname(sysconfig.get_path("include")),]
 library_dirs = [get_python_lib()]
 
+if nvtx_dir := os.getenv("NVTX_DIR"):
+    include_dirs.insert(0, os.path.join(nvtx_dir, "include"))
```
gedoensmax: Actually for most people this will be CUDA_PATH on windows. Did you on purpose name it differently ? 
`C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v11.8\include\nvtx3`

shwina: You're right -- I've changed this to just `NVTX_INCLUDE_DIR` (the path to the directory containing NVTX headers like `nvtoolsExt.h`). So windows users would set it to e.g., `C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v11.8\include\nvtx3`. 

Does that sound reasonable?

gedoensmax: I am also fine doing that ! 👍

jcohen-nvidia: Where is this env var supposed to be set?  Also, I generally don't like to set the include path to .../include/nvtx3, but rather just to .../include, and make the source files do `#include <nvtx3/nvToolsExt.h>` so they don't inadvertently pick up the old version of nvToolsExt.h from the CUDA include directory (where cuda.h is).

shwina: Thanks, @jcohen-nvidia ! I realized that within the Cython source, we do `#include <nvtx3/nvToolsExt.h>`, so the correct include dir is indeed the directory containing `nvtx3/`.

I've changed the variable name (again) to `NVTX_PREFIX`. I've also added some documentation describing how the environment variable should be set, and the corresponding `pip` command that will install `nvtx` from source.

(I also tested locally to confirm that everything works as expected).

PR #72, opened 2023-07-19
Update setup.py to include headers from CUDA_HOME

ppwwyyxx:
NVTX needs headers `nvtx3/nvToolsExt.h` to build. These headers are from the cuda installation.

On systems where cuda is not installed at a default prefix, `CUDA_HOME` is an environment variable that many software (e.g. pytorch) uses to identify where cuda is installed.  https://www.google.com/search?q=CUDA_HOME

This PR adds support so that NVTX can find cuda from the `CUDA_HOME` variable.

### DIFF
--- a/python/setup.py
+++ b/python/setup.py
@@ -22,7 +22,7 @@ except Exception:
 
 include_dirs = [os.path.dirname(sysconfig.get_path("include")),]
 if os.getenv("CUDA_HOME"):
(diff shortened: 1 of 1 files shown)


shwina (approved):
@bdice could you also perhaps take a look?

bdice (approved):
This is fine with me -- to be clear, `CUDA_HOME` is not an official standard but it is in common practice. Thanks for the ping @shwina!

PR #74, opened 2023-07-19
Remove action-checkout-pr as it is not in the allowlist

shwina:


shwina:
@jcohen-nvidia merging this one too as it's needed to unblock CI

PR #76, opened 2023-08-24
Add Python 3.11.

bdice:
This PR adds Python 3.11 support. I changed the build/skip logic so that future CPython versions can be added by just updating the cibuildwheel action -- and versions that are no longer supported can be added to the skip list.

https://cibuildwheel.readthedocs.io/en/stable/options/#build-skip

Since Python 3.12.0rc1 is out, and is guaranteed to be ABI-stable with the final 3.12.0, the latest cibuildwheel version also adds support for that. I believe this PR will also build Python 3.12 as a result.

### DIFF
--- a/.github/workflows/cibw.yml
+++ b/.github/workflows/cibw.yml
@@ -62,7 +62,7 @@ jobs:
         for pyver in 3.8 3.9 3.10 3.11; do
           docker run --rm \
             -v $(pwd)/dist:/wheels \
-            ${{ matrix.docker_py_container}}:$pyver-bullseye \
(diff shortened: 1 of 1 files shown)


bdice:
@shwina Can you please take a look?

shwina (approved):

BwL1289:
Has 3.12 been confirmed to work?

bdice:
@BwL1289 Yes, 3.12 support shipped with 0.2.10 and is in active use by RAPIDS and other libraries. https://pypi.org/project/nvtx/0.2.10/#files

BwL1289:
@bdice thank you!

PR #84, opened 2023-11-30
Fix Windows build when WIN32_LEAN_AND_MEAN is defined by always including stdlib.h (nvbug 200683304)

NBickford-NV:
This fixes the bug where this code:

```cpp
#define WIN32_LEAN_AND_MEAN
#include <nvtx3/nvToolsExt.h>

int main() {}
```

outputs

```output
1>...\include\nvtx3\nvtxDetail\nvtxInit.h(164,32): error C3861: '_wgetenv': identifier not found
```

Normally, `#include <Windows.h>` in nvtxImpl.h would include `<stdlib.h>` — unless `WIN32_LEAN_AND_MEAN` is defined. In that case, nothing defines `_wgetenv`. This pull request should fix that by moving `#include <stdlib.h>` out of the Linux-only preprocessor branch.

Thanks!

### DIFF
--- a/c/include/nvtx3/nvtxDetail/nvtxImpl.h
+++ b/c/include/nvtx3/nvtxDetail/nvtxImpl.h
@@ -12,6 +12,8 @@
 
 /* ---- Include required platform headers ---- */
 
+#include <stdlib.h>
+
 #if defined(_WIN32) 
 
 #include <Windows.h>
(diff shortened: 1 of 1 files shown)


evanramos-nvidia:
Closing this PR because the issue has been fixed on the dev branch.

PR #86, opened 2023-11-30
Python: fix automatic annotation based on function name

### LINKED ISSUE #85: [python] Automatic annotation with function name
Hello,

As of `nvtx 0.2.8` (installed from Pypi), the feature automatically annotating functions with their name when using `@nvtx.annotate()` no longer works.

I believe this is due to the following check:
https://github.com/NVIDIA/NVTX/blob/fcc56f8d439ab33249ab1d702c7daa6e73a353d7/python/nvtx/nvtx.py#L107

Indeed, `self.attributes.message` seems to be created regardless of whether `message` was provided in the constructor:
https://github.com/NVIDIA/NVTX/blob/fcc56f8d439ab33249ab1d702c7daa6e73a353d7/python/nvtx/nvtx.py#L78

Changing the check to:
```py
if not self.attributes.message.string:
```
locally seems to give back the expected behavior (however I don't know if this is the correct fix).

merlinND:
Fixes #85.

Note that I have only tested this locally on my limited use-case.

### DIFF
--- a/python/nvtx/nvtx.py
+++ b/python/nvtx/nvtx.py
@@ -104,7 +104,7 @@ class annotate:
         return False
 
     def __call__(self, func):
-        if not self.attributes.message:
+        if not self.attributes.message.string:
             self.attributes.message = RegisteredString(
                 self.domain.handle, func.__name__
             )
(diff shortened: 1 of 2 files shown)


shwina:
Thanks , @merlinND! This looks good to me. I added a unit test.

shwina:
I'll cut a new release that includes this change today

merlinND:
Hello @shwina,

When you get the chance, could you please publish 0.2.9 to pypi?

PR #88, opened 2024-02-02
Add missing brackets around object construction

r-barnes:
Fixes:
```
nvtx3/nvtx3.hpp:1931:9: error: suggest braces around initialization of subobject [-Werror,-Wmissing-braces]
        0,                              // payload value (union)
        ^
        {}
nvtx3/nvtx3.hpp:1933:9: error: suggest braces around initialization of subobject [-Werror,-Wmissing-braces]
        0                               // message value (union)
        ^
        {}
```
after this the code should be able to compile with `-Wmissing-braces`.


evanramos-nvidia:
Fixed on dev branch by 81b27a306ad9f286349779617fd169f059289645

PR #92, opened 2024-04-25
Fix clang warning on missing braces

bernhardmgruber:
When clang++ is used as host compiler after nvcc, it emits a warning when compiling with `-Wmissing-braces`, because nvcc emits parentheses around the initializer for a union instead of braces.

Warning observed with clang++-17 and nvcc 12.4 when `nvtx3.hpp` is used inside [CCCL](https://github.com/NVIDIA/cccl).


bernhardmgruber:
Seems I was not the first one: #88.

evanramos-nvidia:
Fixed by 81b27a306ad9f286349779617fd169f059289645

PR #100, opened 2024-07-08
Update the headers of the counters extension

rdietrich-nvidia:
* Enable feed-forward (static) counter IDs
* Added a description field to NVTX counters attributes
* Move NVTX counters semantics into a separate header
* Replaced the counter semantics "TIMERANGE" flags with "INTERPOLATION" flags
* Changed "SINCE_START" from being a time scope into an interpolation type.
* Added a definition for the counters semantic version
* Changed the min/max limits union field into a separate field for min and max to avoid ambiguity.
* Changed several comments in the NVTX payload extension header.


c/include/nvtx3/nvToolsExtCounters.h
```diff
@@ -34,46 +27,14 @@
...
-
-/** Counter visualization hints **/
-#define NVTX_COUNTERS_FLAG_INTERPOLATE      (1 << 14)
-
-/** Datatypes for limits union (value of `limitType`). */
-#define NVTX_COUNTERS_LIMIT_I64 0
-#define NVTX_COUNTERS_LIMIT_U64 1
-#define NVTX_COUNTERS_LIMIT_F64 2
+/** Dynamically (tool) generated counter (group) IDs */
+#define NVTX_COUNTERS_ID_DYNAMIC_START 4294967296  /* 1 << 32 */
```
evanramos-nvidia: If this works as-is, great. I get the feeling that it may benefit from adding the literal suffix `ULL` (for `unsigned long long`) which is very likely to be a 64-bit type.

evanramos-nvidia (commented):
Looks good to me, but this should get another review.

c/include/nvtx3/nvToolsExtCounters.h
jcohen-nvidia: We want to stay C89 compatible, and the "LL" suffix was a C99 addition.  The stdint.h types are C99 too, technically, but only very few compilers whine about that in -std=c89 mode, and only with `-Wall` or -`Wlong-long`.  According to the cppreference page for C integer literals, "The type of the integer constant is the first type in which the value can fit".  So this will be parsed as a signed 64-bit integer, which can then undergo implicit promotion to an unsigned 64-bit integer, regardless of whether "long" or "long long" is used for those types.  It would be assignable directly to int64_t and uint64_t.  I tried with a bunch of compilers on godbolt in C89 mode, and I could never get any of them to give me a warning about comparing a signed value with an unsigned value:  https://godbolt.org/z/8G1435rzd

I did however find that some 32-bit GCCs didn't work with this (`error: integer constant is too large for 'long' type`), but it works if the define is `((uint64_t)1 << 32)` instead of 4294967296.  Since all other compilers liked the expression (assuming they weren't whining about stdint.h), I think that reads a bit better and is closer to our existing pattern, so I think we should change to this.   Note the `32` will be an `int` literal, but the `<<` operator will promote it to the same type as the other argument -- in cases like this where I have two `int` literals and want to promote to 64-bit/float/double before doing the operator, I like to cast or use a suffix on only the first operand and let the other promote.

rdietrich-nvidia: I changed it to `((uint64_t)1 << 32)`, also in the payloads header.

evanramos-nvidia (approved):

PR #103, opened 2024-09-10
Fix macOS build

evanramos-nvidia:


jcohen-nvidia:
Pretty unfortunate that Apple doesn't ship uchar.h and have char16/32 support, even though they claim to on cppreference:

image

But it seems other people have hit this as well:
https://stackoverflow.com/questions/24608832/uchar-h-file-not-found-on-os-x-10-9

I think your fix here makes sense, given there doesn't seem to be any version of the Apple platform that supports these types.

jcohen-nvidia (approved):

PR #114, opened 2025-02-23
Make NVTX CMake relocatable.

### LINKED ISSUE #113: NVTX should ship CMake package config so that find_package works
Currently it requires a lot of hacks and non-obvious knowledge to use packages of NVTX with CMake, such as `nvtx-c` on `conda-forge`, because the library does not ship a proper package config file.

https://github.com/conda-forge/pytorch-cpu-feedstock/issues/357 is a good example of the kind of problem this creates -- the `nvtx-c` packages contain NVTX headers but they are not findable by the build system. Several packaging experts and maintainers (including me) were unable to point to a correct solution.

Presently, conda packages that want to get NVTX as a dependency must install `nvtx-c` _and_ tell their build system something like "trust me, it's already on the include path." This also makes it very hard to support CPM-based workflows that try to find the package locally and fetch it from upstream if not found locally (the find-local step always fails).

As a workaround for this problem, RAPIDS [repackages NVTX with package exports and other configuration](https://github.com/rapidsai/rapids-cmake/blob/branch-25.04/rapids-cmake/cpm/nvtx3.cmake) so that it includes the following files:
```
lib/cmake/nvtx3/nvtx3-config-version.cmake
lib/cmake/nvtx3/nvtx3-config.cmake
lib/cmake/nvtx3/nvtx3-targets.cmake
```

Currently `librmm` conda packages vendor nvtx3, but only because I can't make `nvtx-c` work without additional package configuration.

I think the solution is that we need to ship `lib/cmake/nvtx3/nvtx3-config.cmake` as a part of the NVTX tarballs published to GitHub, but I would like confirmation from others.

cc: @robertmaynard @KylefromNVIDIA @vyasr @evanramos-nvidia for awareness

bdice:
This resolves #113.

I added a config and targets file, using the existing logic in `CMakeLists.txt` and `nvtxImportedTargets.cmake`. I tried to modify these files in a way that wouldn't break existing uses of the CMake code (which apparently only works via `add_subdirectory`).

I tested these changes with:

```bash
mkdir -p /tmp/nvtx-install
mkdir -p c/build

pushd c/build
rm -rf * && rm -rf /tmp/nvtx-install/* && cmake .. -DCMAKE_INSTALL_PREFIX=/tmp/nvtx-install && cmake --build . && cmake --install .
popd

mkdir -p c/examples/build

pushd c/examples/build
rm -rf * && cmake .. -DCMAKE_PREFIX_PATH=/tmp/nvtx-install && cmake --build .
popd
```

I think the NVTX relocatable install in `/tmp/nvtx-install` looks like it should work, based on my knowledge of how rapids-cmake works (https://github.com/rapidsai/rapids-cmake/blob/branch-25.04/rapids-cmake/cpm/nvtx3.cmake). The example code builds and runs with the appropriate include directories.

To move this PR forward, this may need some or all of the following:
- Better CMake test code in CI, perhaps this can be done in a separate PR prior to merging this PR. The current code in `c/examples` should probably be in a directory of CMake tests somewhere.
- Some kind of documentation changes? The NVTX docs appear to live in several places.
- Review from two of the RAPIDS CMake folks: @robertmaynard, @vyasr, @KylefromNVIDIA
- Review from NVTX maintainers

### DIFF
--- a/c/CMakeLists.txt
+++ b/c/CMakeLists.txt
@@ -36,47 +36,47 @@ cmake_minimum_required(VERSION 3.10)
 # Project declaration needs to be before the configuration
 project(nvtx3 VERSION 3.1.1)
 
-# Add installation rules
-include(GNUInstallDirs)
-include(CMakePackageConfigHelpers)
+# Determine if this is the root project
+set(NVTX3_IS_ROOT_PROJECT "${CMAKE_SOURCE_DIR}" STREQUAL "${PROJECT_SOURCE_DIR}")
+option(NVTX3_INSTALL "Install NVTX3" ${NVTX3_IS_ROOT_PROJECT})
 
 set(NVTX3_TARGETS_NOT_USING_IMPORTED ON)
 include(nvtxImportedTargets.cmake)
+set_target_properties(nvtx3-c PROPERTIES EXPORT_PROPERTIES VERSION)
 
+if(NVTX3_INSTALL)
+    # Add installation rules
+    include(GNUInstallDirs)
+    include(CMakePackageConfigHelpers)
 
-set(INCLUDE_INSTALL_DIR "${CMAKE_INSTALL_INCLUDEDIR}/nvtx3")
+    # Configure the config file
+    configure_package_config_file(
+        "${CMAKE_CURRENT_SOURCE_DIR}/nvtx3-config.cmake.in"
+        "${CMAKE_CURRENT_BINARY_DIR}/nvtx3-config.cmake"
+        INSTALL_DESTINATION "${CMAKE_INSTALL_LIBDIR}/cmake/nvtx3"
+    )
 
-# Configure the config file
-configure_package_config_file(
-    "${CMAKE_CURRENT_SOURCE_DIR}/nvtx3-config.cmake.in"
-    "${CMAKE_CURRENT_BINARY_DIR}/nvtx3-config.cmake"
-    INSTALL_DESTINATION "${CMAKE_INSTALL_LIBDIR}/cmake/nvtx3"
-    PATH_VARS INCLUDE_INSTALL_DIR
-)
+    # Generate version file
+    write_basic_package_version_file(
+        "${CMAKE_CURRENT_BINARY_DIR}/nvtx3-config-version.cmake"
+        VERSION ${PROJECT_VERSION}
+        COMPATIBILITY SameMajorVersion
+    )
 
-# Generate version file
-write_basic_package_version_file(
-    "${CMAKE_CURRENT_BINARY_DIR}/nvtx3-config-version.cmake"
-    VERSION ${PROJECT_VERSION}
-    COMPATIBILITY SameMajorVersion
-)
+    # Install headers
+    install(DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/include/"
+            DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}")
 
-# Install headers
-install(DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/include/"
-        DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}"
-        FILES_MATCHING PATTERN "*.h"
-                       PATTERN "*.hpp")
+    # Install the config files
+    install(FILES
+        "${CMAKE_CURRENT_BINARY_DIR}/nvtx3-config.cmake"
+        "${CMAKE_CURRENT_BINARY_DIR}/nvtx3-config-version.cmake"
+        DESTINATION "${CMAKE_INSTALL_LIBDIR}/cmake/nvtx3"
+    )
 
-# Install the config files
-install(FILES
-    "${CMAKE_CURRENT_BINARY_DIR}/nvtx3-config.cmake"
-    "${CMAKE_CURRENT_BINARY_DIR}/nvtx3-config-version.cmake"
-    DESTINATION "${CMAKE_INSTALL_LIBDIR}/cmake/nvtx3"
-)
-
-# Install the targets
-install(EXPORT nvtx3-targets
-    FILE nvtx3-targets.cmake
-    NAMESPACE nvtx3::
-    DESTINATION "${CMAKE_INSTALL_LIBDIR}/cmake/nvtx3")
-
+    # Install the targets
+    install(EXPORT nvtx3-targets
+        FILE nvtx3-targets.cmake
+        NAMESPACE nvtx3::
+        DESTINATION "${CMAKE_INSTALL_LIBDIR}/cmake/nvtx3")
+endif()
--- a/c/CMakeLists.txt
+++ b/c/CMakeLists.txt
@@ -48,6 +48,7 @@ if(NVTX3_INSTALL)
     include(CMakePackageConfigHelpers)
 
     # Configure the config file
+    message(STATUS "Writing config file to ${CMAKE_CURRENT_BINARY_DIR}/nvtx3-config.cmake")
     configure_package_config_file(
         "${CMAKE_CURRENT_SOURCE_DIR}/nvtx3-config.cmake.in"
         "${CMAKE_CURRENT_BINARY_DIR}/nvtx3-config.cmake"
--- a/c/CMakeLists.txt
+++ b/c/CMakeLists.txt
@@ -33,5 +33,50 @@ cmake_minimum_required(VERSION 3.10)
 # IMPORTED option, which only defines the targets within the scope of the library.
 # This allows the same target name to be used for different NVTX versions.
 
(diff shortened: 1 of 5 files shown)


c/CMakeLists.txt
```diff
@@ -33,5 +33,50 @@ cmake_minimum_required(VERSION 3.10)
...
+write_basic_package_version_file(
+    "${CMAKE_CURRENT_BINARY_DIR}/nvtx3-config-version.cmake"
+    VERSION ${PROJECT_VERSION}
+    COMPATIBILITY SameMajorVersion
+)
+
+# Install headers
+install(DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/include/"
+        DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}"
+        FILES_MATCHING PATTERN "*.h"
```
robertmaynard: There is no need for the FILES_MATCHING

c/CMakeLists.txt
```diff
@@ -33,5 +33,50 @@ cmake_minimum_required(VERSION 3.10)
...
+# Generate version file
+write_basic_package_version_file(
+    "${CMAKE_CURRENT_BINARY_DIR}/nvtx3-config-version.cmake"
+    VERSION ${PROJECT_VERSION}
+    COMPATIBILITY SameMajorVersion
+)
+
+# Install headers
+install(DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/include/"
+        DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}"
```
robertmaynard: We should comment on why we aren't using `INCLUDE_INSTALL_DIR` here

c/CMakeLists.txt
```diff
@@ -33,5 +33,50 @@ cmake_minimum_required(VERSION 3.10)
...
+# Project declaration needs to be before the configuration
+project(nvtx3 VERSION 3.1.1)
+
+# Add installation rules
+include(GNUInstallDirs)
+include(CMakePackageConfigHelpers)
+
 set(NVTX3_TARGETS_NOT_USING_IMPORTED ON)
 include(nvtxImportedTargets.cmake)
+
```
robertmaynard, line 44: You need to add `set_target_properties(nvtx3-c PROPERTIES EXPORT_PROPERTIES VERSION)` after the include so that we match the version property checks in the `nvtxImportedTargets.cmake`

c/CMakeLists.txt
```diff
@@ -33,5 +33,50 @@ cmake_minimum_required(VERSION 3.10)
...
+
+set(INCLUDE_INSTALL_DIR "${CMAKE_INSTALL_INCLUDEDIR}/nvtx3")
```
robertmaynard: All the install rules should be guarded behind a check that looks like:
```cmake
set(NVTX3_IS_ROOT_PROJECT .....)
option(NVTX3_INSTALL ${NVTX3_IS_ROOT_PROJECT})
if(NVTX3_INSTALL)
....
endif()
```

That way existing projects don't start installing nvtx3 where previously they didn't.

robertmaynard (changes requested):

c/CMakeLists.txt
bdice: I don't think I need `INCLUDE_INSTALL_DIR` anymore, it wasn't being used in `nvtx3-config.cmake.in` so I got rid of it. Is that an equally good solution?

c/CMakeLists.txt
bdice: Removed.

c/CMakeLists.txt
bdice: I added this for `nvtx3-c` and also `nvtx3-cpp` -- does it look right now? I am not 100% sure on this.

c/CMakeLists.txt
bdice: I guessed that the `.....` was supposed to be this condition:

```cmake
set(NVTX3_IS_ROOT_PROJECT "${CMAKE_SOURCE_DIR}" STREQUAL "${PROJECT_SOURCE_DIR}")
```

If that's correct, I think this is fixed.

robertmaynard: You can use `PROJECT_IS_TOP_LEVEL` https://cmake.org/cmake/help/latest/variable/PROJECT_IS_TOP_LEVEL.html

If we can rely on CMake 3.21 otherwise you generally check `CMAKE_PROJECT_NAME` matches our project name ( `PROJECT_NAME` )

c/CMakeLists.txt
robertmaynard: Looks great

c/CMakeLists.txt
robertmaynard: Looks great

evanramos-nvidia:
@bdice: Thanks for taking this on.

@robertmaynard: Thanks for giving feedback.

We may need to adjust `tests/CMakeLists.txt` based on your improvements so that our CI passes. I'm not sure if there are ways that file and directory should make more idiomatic use of CMake. The one thing I want to avoid is users building the tests by mistake when they just want to use NVTX in their own project.

c/CMakeLists.txt
```diff
@@ -33,5 +33,50 @@ cmake_minimum_required(VERSION 3.10)
 # IMPORTED option, which only defines the targets within the scope of the library.
 # This allows the same target name to be used for different NVTX versions.
 
+# Project declaration needs to be before the configuration
+project(nvtx3 VERSION 3.1.1)
+
+# Determine if this is the root project
+set(NVTX3_IS_ROOT_PROJECT "${CMAKE_SOURCE_DIR}" STREQUAL "${PROJECT_SOURCE_DIR}")
```
dominik-wojt-2311446: I am afraid cmake will not accept the `if` syntax with `set`. At least version 3.22.1 I tried it with did not.
This fixed the issue at my end:
```
if("${CMAKE_SOURCE_DIR}" STREQUAL "${PROJECT_SOURCE_DIR}")
    set(NVTX3_IS_ROOT_PROJECT TRUE)
else()
    set(NVTX3_IS_ROOT_PROJECT FALSE)
endif()
```

dominik-wojt-2311446 (commented):
I was half way trying to do the same at my end. The lack of proper cmake scripts for `find_package` is really annoying.

Your work is really appreciated.

You might want to add this snippet to `README.md`:
````
## Use installed version of NVTX

If NVTX is installed in your system, you can use it with a standard `find_package` procedure.

```cmake
find_package(nvtx3 REQUIRED)

add_executable(some_c_program main.c)
target_link_libraries(some_c_program PRIVATE nvtx3::nvtx3-c)

add_executable(some_cpp_program main.cpp)
target_link_libraries(some_cpp_program PRIVATE nvtx3::nvtx3-cpp)
```
````

File: c/nvtxImportedTargets.cmake [hunk 6]
```diff
@@ -55,18 +55,23 @@ else()
...
     #-------------------------------------------------------
     # Define "nvtx3-cpp" library for the NVTX v3 C++ API
     # Separate target allows attaching independent compiler requirements if needed
     add_library(nvtx3-cpp INTERFACE ${OPTIONALLY_IMPORTED})
+    add_library(nvtx3::nvtx3-cpp ALIAS nvtx3-cpp)
     set_target_properties(nvtx3-cpp PROPERTIES VERSION ${NVTX3_VERSION})
     target_link_libraries(nvtx3-cpp INTERFACE nvtx3-c)
+
+    # Install the targets
+    install(TARGETS nvtx3-c nvtx3-cpp EXPORT nvtx3-targets)
```
dominik-wojt-2311446: I think this line should be moved to `CMakeLists.txt` and protected with `NVTX3_INSTALL`.
It fails when `OPTIONALLY_IMPORTED` is `IMPORTED`.

c/CMakeLists.txt
robertmaynard: You are correct. CMake doesn't support the original syntax.

I would use:
```cmake
set(NVTX3_IS_ROOT_PROJECT FALSE)
if("${CMAKE_SOURCE_DIR}" STREQUAL "${PROJECT_SOURCE_DIR}")
    set(NVTX3_IS_ROOT_PROJECT TRUE)
endif()
```

BwL1289:
@bdice thank you for working on this. We currently can't properly use nvtx in downstream packages (like torch) without this PR.

bdice:
@dominik-wojt-2311446 @BwL1289 Thanks! I will finish this up shortly.

bdice (commented):
@KyleFromNVIDIA Thanks so much for your help. Your changes look good to me.

@evanramos-nvidia Could you help review this?

KyleFromNVIDIA:
This PR has been demonstrated to work by https://github.com/rapidsai/rapids-cmake/pull/804.

PR #122, opened 2025-07-01
Python fix: Make @nvtx.annotate close the NVTX range when an exception is thrown

paultill:
I was using nvtx.annotate as a Python decorator like this:

```
@nvtx.annotate("foo")
def some_function():
      # code throws exception some time.
```

I found that if some_function threw an exception, the NVTX range "foo" was never closed, so its duration would seem to extend all the way to end of the trace, even though the some_function is no longer executing.

By contrast, using nvtx.annotate as a context manager works as expected, which is that even when an exception is thrown, the NVTX range is closed as soon as the block of code is exited:

```
with nvtx.annotate("foo"):
   # code throws exception
```

I made a change to ensure the that first case behaves like the second, so any NVTX range opened by an annotated function will be closed at the time that function exits by throwing an exception.

I also added tests to specify the expected behavior.

### DIFF
--- a/python/tests/test_nvtx_annotate_pushes_and_pops.py
+++ b/python/tests/test_nvtx_annotate_pushes_and_pops.py
@@ -1,0 +1,141 @@
+# SPDX-FileCopyrightText: Copyright (c) 2020-2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://nvidia.github.io/NVTX/LICENSE.txt for license information.
+
+import unittest.mock
+
+import pytest
+
+import nvtx
+
+
+def test_annotate_ctx_manager_pops():
+    """
+    Test that the NVTX range is properly popped when the context manager is exited normally.
+    """
+    # Create a mock domain that behaves like a real domain
+    mock_domain = unittest.mock.MagicMock()
+    mock_domain.handle = unittest.mock.MagicMock()
+    mock_domain.get_event_attributes.return_value = unittest.mock.MagicMock()
+    mock_domain.get_registered_string.return_value = unittest.mock.MagicMock()
+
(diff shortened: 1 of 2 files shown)


evanramos-nvidia:
Thanks for the observation and the patch! We'll review it shortly.

python/tests/test_nvtx_annotate_pushes_and_pops.py
```diff
@@ -0,0 +1,141 @@
...
+    mock_domain.get_registered_string.return_value = unittest.mock.MagicMock()
+
+    # Mock get_domain to return our mock domain instead of dummy_domain
+    with unittest.mock.patch(
+        "nvtx.nvtx.get_domain", return_value=mock_domain
+    ), unittest.mock.patch(
+        "nvtx.nvtx.libnvtx_push_range"
+    ) as mock_push, unittest.mock.patch(
+        "nvtx.nvtx.libnvtx_pop_range"
+    ) as mock_pop:
```
guysz-nvidia: This code block is duplicated on each test.
I suggest defining the mocks as fixtures for eliminating code duplication and for having a better separation between setup phase and test phase.

Here is what I had in mind:
```python
@pytest.fixture
def mock_domain():
    mock_domain = unittest.mock.MagicMock()
    mock_domain.handle = unittest.mock.MagicMock()
    mock_domain.get_event_attributes.return_value = unittest.mock.MagicMock()
    mock_domain.get_registered_string.return_value = unittest.mock.MagicMock()
    with unittest.mock.patch("nvtx.nvtx.get_domain", return_value=mock_domain) as mock:
        yield mock

@pytest.fixture
def mock_push():
    with unittest.mock.patch("nvtx.nvtx.libnvtx_push_range") as mock:
        yield mock

@pytest.fixture
def mock_pop():
    with unittest.mock.patch("nvtx.nvtx.libnvtx_pop_range") as mock:
        yield mock
    
def test_annotate_ctx_manager_pops(mock_domain, mock_push, mock_pop):
    with nvtx.annotate():
        pass

    mock_push.assert_called_once()
    mock_pop.assert_called_once()


# And so on with the other tests
```

guysz-nvidia (changes requested):
The suggested fix looks good to me.
Posted a comment regarding the test implementation.

Thank you for adding this!

python/tests/cython/test_profiler.c
guysz-nvidia, line 1: Please remove this file from the change list, as it is generated by Cython and should not be tracked by git.

guysz-nvidia (changes requested):

python/tests/cython/test_profiler.c
paultill: Oops, just noticed that. I removed it.

guysz-nvidia (approved):
The changes look good to me.
Thank you!

paultill:
Thanks for the review! 

Do you need anything else from me? 

(Will someone else merge this change?)

guysz-nvidia:
> Thanks for the review!
> 
> Do you need anything else from me?
> 
> (Will someone else merge this change?)

The changes look good from my side. Should be merged by @evanramos-nvidia

paultill:
@evanramos-nvidia Do you need anything else from me before you merge?

